PR #9, opened 2018-10-27
Fixes to class comparisons and ordering for Python 3 compatibility

llimeht:
Two changes included here for Python 3 compatibility:
  * String comparisons with `__class__` are the wrong way to look at types in Python 3, `isinstance` is better
  * `cmp` no longer exists in Python 3 and for sorting, comparison functions such as `__gt__` and `__eq__` must be given instead

### DIFF
--- a/modules/pymol/editing.py
+++ b/modules/pymol/editing.py
@@ -24,6 +24,7 @@ if __name__=='pymol.editing':
           boolean_sc,boolean_dict,safe_list_eval, is_sequence, \
           DEFAULT_ERROR, DEFAULT_SUCCESS, _raising, is_ok, is_error              
     from chempy import cpv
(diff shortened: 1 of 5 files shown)


modules/pymol/editing.py
```diff
@@ -24,6 +24,7 @@
           boolean_sc,boolean_dict,safe_list_eval, is_sequence, \
           DEFAULT_ERROR, DEFAULT_SUCCESS, _raising, is_ok, is_error              
     from chempy import cpv
+    import pymol.wizard.dragging
```
llimeht: There's a chance that this now creates a circular import with Python 2 -- @speleo3 is there a CI test suite that can help catch these things?

PR #12, opened 2018-11-16
Document the overlap command

pslacerda:
I needed an overlap command that don't be larger if there are more atoms, so I had to look for what exactly `cmd.overlap` does, and it values is larger for bigger selections. Just documented it.


speleo3:
thank you!

PR #34, opened 2019-06-13
Update managergui_qt.py

carlosfamilia:
Correct QtListWidgets method call from selecteditems to selectedItems.


speleo3:
thanks for the fix!

carlosfamilia:
No problem.
Always glad to help.

PR #45, opened 2019-07-03
Support fetching two letter chains

pslacerda:
The structure 4WQF has chains with two letter codes. This commit allow `fetch` to import such chains (eg. `fetch 4WQF.BZ`).

### DIFF
--- a/modules/pymol/importing.py
+++ b/modules/pymol/importing.py
@@ -1313,8 +1313,15 @@ PYMOL API
                     obj_name = 'emd_' + obj_code
 
             chain = None
-            if len(obj_code) in (5,6) and type in ('pdb', 'cif', 'mmtf'):
-                obj_code, chain = obj_code[:4], obj_code[-1]
+            if len(obj_code) in (5,6,7) and type in ('pdb', 'cif', 'mmtf'):
+                obj_code = (
+                    obj_code
+                        .replace('.', '')
+                        .replace('_', '')
+                        .replace('-', '')
+                        .replace(':', '')
+                    )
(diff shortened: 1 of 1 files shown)


speleo3:
Looks fine. One question: is `.` a common separator for PDB code and chain? I have seen others (like `:` or `_`) but not `.`.

pslacerda:
Just added these separators, seems to work fine. Yes, `.` it is valid, although  is a bit annoying because the newly created object name isn't standardized (ie. converted to `_` as the others).

speleo3:
Thanks for the update. I think we should do it like this:

```python
            chain = None
            if type in ('pdb', 'cif', 'mmtf'):
                m = re.match(r'(\d[0-9a-z]{3})[:_.]?([a-z]+)$', obj_code, flags=re.I)
                if m is not None:
                    obj_code, chain = m.group(1), m.group(2)
```

pslacerda:
Never know when 5 character PDB ids will come along.

speleo3:
8 character PDB codes are already there:
https://www.rcsb.org/news?year=2017&article=59cd3fa822e913222ac88fcd

pslacerda:
Nice update.

> 8 character PDB codes are already there:
> https://www.rcsb.org/news?year=2017&article=59cd3fa822e913222ac88fcd

There is any id that was already replaced by the new standard? If the updates happens as the article said, old accession codes will increasingly be obsoleted.

speleo3:
> There is any id that was already replaced by the new standard?

I don't think so.

PR #64, opened 2019-11-10
cmd.dump quiet flag and documentation

### LINKED ISSUE #62: dump verbose option
Hi,

The dump option could have a verbose option.

Seems that ExecutiveDump, ObjectMeshDump and ObjectSurfaceDump should be modified to have such option.

pslacerda:
I copied the documentation from https://pymolwiki.org/index.php/Dump, is it the right approach?

Closes #62

### DIFF
--- a/modules/pymol/experimenting.py
+++ b/modules/pymol/experimenting.py
@@ -133,18 +133,54 @@ DESCRIPTION
             print(" minimize: missing parameters, can't continue")
 
 
-    def dump(fnam,obj,_self=cmd):
+    def dump(fnam, obj, quiet=1, _self=cmd):
         '''
 DESCRIPTION
 
-    "dump" is an unsupported command which may have something to do
-    with outputing isosurface meshes and surface objects to a file.
+    The dump command writes the geometry of an isosurface, isomesh,
+    or isodot object to a simple text file. Each line contains one
+    vertex.
+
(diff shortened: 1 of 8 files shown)


modules/pymol/experimenting.py
```diff
@@ -133,18 +133,54 @@ def minimize(sele='', iter=500, grad=0.01, interval=50, _setup=1, _self=cmd):
             print(" minimize: missing parameters, can't continue")
 
 
-    def dump(fnam,obj,_self=cmd):
+    def dump(fnam, obj, state=0, quiet=1, _self=cmd):
```
speleo3: PyMOL's state handling is 1-based in the Python API, and 0-based in the C-API. So it should be `state=1` here...

modules/pymol/experimenting.py
```diff
@@ -133,18 +133,54 @@ def minimize(sele='', iter=500, grad=0.01, interval=50, _setup=1, _self=cmd):
...
+SEE ALSO
+
+    COLLADA export
 
     '''
         r = DEFAULT_ERROR
         try:
             _self.lock(_self)
-            r = _cmd.dump(_self._COb,str(fnam),obj)
+            r = _cmd.dump(_self._COb, str(fnam), obj, int(state), int(quiet))
```
speleo3: ... and `int(state) - 1` here

PR #65, opened 2019-11-12
Add exec call to launcher

pslacerda:
Is more sane and is a lot less trickier to embed the PyMOL window inside another program because the window id is related to PID directly as there is no sh process leftover anymore.


speleo3:
Thanks! I've rebased this commit, please make sure to branch from latest master before you make your next pull request :-)

PR #67, opened 2019-11-16
Export maps with cmd.dump

pslacerda:
Hi,

Currently I'm `cmd.dump`ing isodots level by level until have a filled volume. But with these changes I can get the original grid map which is better.

Unfortunately I was confused about the ttt matrix and don't know how to transform coordinates.

For clearance floats with sizeof 2 was ignored and I have no way to test data grids, only float.

### DIFF
--- a/layer2/ObjectMap.cpp
+++ b/layer2/ObjectMap.cpp
@@ -5979,20 +5979,19 @@ void ObjectMapDump(ObjectMap* om, const char* fname, int state, int quiet)
   for (int xi = 0; xi < oms->Field->dimensions[0]; xi++) {
     for (int yi = 0; yi < oms->Field->dimensions[1]; yi++) {
       for (int zi = 0; zi < oms->Field->dimensions[2]; zi++) {
-        float x = oms->ExtentMin[0] + xs * xi;
-        float y = oms->ExtentMin[1] + ys * yi;
-        float z = oms->ExtentMin[2] + zs * zi;
-        
-        CField* field = oms->Field->data;
-        
-        switch (field->type) {
+
+        float x = Ffloat4(oms->Field->points, xi, yi, zi, 0);
+        float y = Ffloat4(oms->Field->points, xi, yi, zi, 1);
+        float z = Ffloat4(oms->Field->points, xi, yi, zi, 2);
+
+        switch (oms->Field->data->type) {
           case cFieldFloat: {
-            float value = Ffloat3(field, xi, yi, zi);
+            float value = Ffloat3(oms->Field->data, xi, yi, zi);
             fprintf(file, "%10.4f%10.4f%10.4f%10.4f\n", x, y, z, value);
             break;
           }
           case cFieldInt: {
-            int value = Fint3(field, xi, yi, zi);
+            int value = Fint3(oms->Field->data, xi, yi, zi);
             fprintf(file, "%10.4f%10.4f%10.4f%10d\n", x, y, z, value);
             break;
           }
--- a/layer2/ObjectMap.cpp
+++ b/layer2/ObjectMap.cpp
@@ -5956,3 +5956,44 @@ ObjectMap *ObjectMapLoadChemPyMap(PyMOLGlobals * G, ObjectMap * I, PyObject * Ma
   return (I);
 #endif
 }
+
+void ObjectMapDump(ObjectMap* om, const char* fname, int state, int quiet)
+{
+  auto file = fopen(fname, "wb");
+  if (!file) {
+    ErrMessage(om->G, "ObjectMapDump", "can't open file for writing");
+  }
+  
+  ObjectMapState* oms = om->State + 0;
+  
+  int zd = oms->Field->dimensions[2];
+  int yd = oms->Field->dimensions[1];
+  int xd = oms->Field->dimensions[0];
+  
+  float zs = (oms->ExtentMax[2] - oms->ExtentMin[2]) / (oms->Field->dimensions[2]);
+  float ys = (oms->ExtentMax[1] - oms->ExtentMin[1]) / (oms->Field->dimensions[1]);
+  float xs = (oms->ExtentMax[0] - oms->ExtentMin[0]) / (oms->Field->dimensions[0]);
+  
+  int offset = 0;
+  for (int zi = 0; zi < oms->Field->dimensions[2]; zi++) {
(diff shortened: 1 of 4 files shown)


layer2/ObjectMap.cpp
```diff
@@ -5956,3 +5956,58 @@ ObjectMap *ObjectMapLoadChemPyMap(PyMOLGlobals * G, ObjectMap * I, PyObject * Ma
...
+  for (int xi = 0; xi < oms->Field->dimensions[0]; xi++) {
+    for (int yi = 0; yi < oms->Field->dimensions[1]; yi++) {
+      for (int zi = 0; zi < oms->Field->dimensions[2]; zi++) {
+        float x = oms->ExtentMin[0] + xs * xi;
+        float y = oms->ExtentMin[1] + ys * yi;
+        float z = oms->ExtentMin[2] + zs * zi;
+        
+        CField* field = oms->Field->data;
+        
+        if(field->type == cFieldFloat) {
```
JarrettSJohnson: Not sure if this if/else block is entirely reliable since the size of types can be different across platforms, and we could incidentally have user-defined types that have the same size as built-in types. As far as I understand, we only have Field data of `int`, `float`, or `Tetsurf::PointType`/`IsoSurf::PointType`. For now, might be best to just query `cFieldFloat`, `cFieldInt`, and probably just ignore the rest. `Field` of `ObjectMapState` seem to only have underlying `float` anyway...

pslacerda: So maybe stick only with `float`? If so then also get rid of `field->type` seems alright.

Would be reliable because importing and exporting already rely on float (eg `ObjectMapXPLORStrToMap`). Also because sessions rely on Python objects that don't have such kind of problems (eg `ObjectMapStateAsPyList`).

The options would be:

```
        switch (field->type) {
          case cFieldFloat: value = (float)((float*)field)[offset]; break;
          case cFieldInt: value = (double)((int*)field)[offset]; break;
          default:
            ErrMessage(om->G, "ObjectMapDump", "unknown field type");
        }
```

or 
 ```
        float value = (float)((float*)field)[offset];
```

Comment (pslacerda) [reply layer2/ObjectMap.cpp]: Maybe I talked nonsense.

File: layer2/ObjectMap.cpp [hunk 2]
```diff
@@ -5956,3 +5956,57 @@ ObjectMap *ObjectMapLoadChemPyMap(PyMOLGlobals * G, ObjectMap * I, PyObject * Ma
...
+      for (int zi = 0; zi < oms->Field->dimensions[2]; zi++) {
+        float x = oms->ExtentMin[0] + xs * xi;
+        float y = oms->ExtentMin[1] + ys * yi;
+        float z = oms->ExtentMin[2] + zs * zi;
+        
+        CField* field = oms->Field->data;
+        
+        switch (field->type) {
+          case cFieldFloat: {
+            float value = ((float*)field->data)[offset];
```
speleo3: Not sure if map data is guaranteed to be contiguous in PyMOL. The `CField` data structure has a `stride` array and macros to look up a data point. I think this will be more correct:
```c++
            float value = Ffloat3(field, xi, yi, zi);
```
```c++
            int value = Fint3(field, xi, yi, zi);
```

pslacerda:
Up this point I just want to get the grid matrix. The mail list says about `get_volume_field` and `get_extents`. Do I need `get_object_matrix`? The commented out lines are required to work properly?

```python
    def get_map_grid(objName, state=1, _self=cmd):
        with _self.lockcm:
            field = _self._cmd.get_volume_field(_self._COb, objName, int(state) - 1, 0)
            extents = _cmd.get_min_max(_self._COb, objName, int(state)-1)
            #trans = _cmd.get_object_matrix(_self._COb, objName, int(state)-1, 1)

        import numpy as np
        import itertools

        #trans = np.array(trans).reshape(4, 4)[:3,:3]

        xs = (extents[1][0] - extents[0][0]) / field.shape[0]
        ys = (extents[1][1] - extents[0][1]) / field.shape[1]
        zs = (extents[1][2] - extents[0][2]) / field.shape[2]

        x = np.arange(field.shape[0]) * xs + extents[0][0]
        y = np.arange(field.shape[1]) * ys + extents[0][1]
        z = np.arange(field.shape[2]) * zs + extents[0][2]

        n = x.shape[0] * y.shape[0] * z.shape[0]
        xyz = np.array(list(itertools.product(x, y, z)))
        #xyz = xyz.dot(trans)

        data = field.reshape((n, 1))
        grid = np.hstack([xyz, data])
        return grid
```

layer2/ObjectMap.cpp
```diff
@@ -5956,3 +5957,56 @@ ObjectMap *ObjectMapLoadChemPyMap(PyMOLGlobals * G, ObjectMap * I, PyObject * Ma
...
+  float ys = (oms->ExtentMax[1] - oms->ExtentMin[1]) / (oms->Field->dimensions[1]);
+  float zs = (oms->ExtentMax[2] - oms->ExtentMin[2]) / (oms->Field->dimensions[2]);
+  
+  int offset = 0;
+  for (int xi = 0; xi < oms->Field->dimensions[0]; xi++) {
+    for (int yi = 0; yi < oms->Field->dimensions[1]; yi++) {
+      for (int zi = 0; zi < oms->Field->dimensions[2]; zi++) {
+        float x = oms->ExtentMin[0] + xs * xi;
+        float y = oms->ExtentMin[1] + ys * yi;
+        float z = oms->ExtentMin[2] + zs * zi;
```
speleo3: I think the `(ExtentMax - ExtentMin) / dimensions` calculation is off by one grid point? Maybe it makes sense to use `Isofield::points` instead.
```c++
  const CField *field = oms->Field->data;
  const CField *points = oms->Field->points;

  for (int xi = 0; xi < oms->Field->dimensions[0]; xi++) {
    for (int yi = 0; yi < oms->Field->dimensions[1]; yi++) {
      for (int zi = 0; zi < oms->Field->dimensions[2]; zi++) {
        float x = F4Ptr(points, xi, yi, zi, 0);
        float y = F4Ptr(points, xi, yi, zi, 1);
        float z = F4Ptr(points, xi, yi, zi, 2);
```

pslacerda:
Seems done.

speleo3:
Thanks Pedro! I've manually merged this with commit bdc9b11de9a58a3c736582086c895b99c7994622

PR #105, opened 2020-06-26
Removed "using namespace std;" in layer[012] files

vedranmiletic:
Small clean-up for consistency with other code.


JarrettSJohnson:
Thanks for the PR.

PR #134, opened 2020-09-28
Include <cstddef> in ov_types.h.

ericonr:
On musl libc, ptrdiff_t, the type used in this header's typedefs, isn't
pulled in by any other include, so it is necessary to include <cstddef>
directly.

### DIFF
--- a/layer0/os_predef.h
+++ b/layer0/os_predef.h
@@ -82,11 +82,7 @@ Z* -------------------------------------------------------------------
 
 /* END PROPRIETARY CODE SEGMENT */
 
-#ifdef __linux__
-#include <malloc.h>
-#else
 #include <stddef.h>
-#endif
 
 #if defined(_MSC_VER)
 // conversion from '...' to '...', possible loss of data
--- a/ov/src/ov_port.h
+++ b/ov/src/ov_port.h
@@ -11,12 +11,7 @@
 #include <string.h>
 #include <stdlib.h>
 #include <ctype.h>
-
-#ifdef __linux__
-#include <malloc.h>
-#else
 #include <stddef.h>
-#endif
 
 #include "ov_defines.h"
 #include "ov_status.h"
--- a/ov/src/ov_types.h
+++ b/ov/src/ov_types.h
@@ -11,6 +11,8 @@
 #ifndef _H_ov_types
 #define _H_ov_types
 
+#include <stddef.h>
+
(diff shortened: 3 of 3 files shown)


speleo3:
Two thoughts:
1) `#include <cstddef>` (C++ only) after `extern "C" {` (indicates a C header) looks wrong. Can you instead use the C version (`stddef.h`) and place it before the `extern "C" {`?
2) I wonder if `stddef.h` should be included for `__linux__` here as well:
https://github.com/schrodinger/pymol-open-source/blob/master/ov/src/ov_port.h#L15
https://github.com/schrodinger/pymol-open-source/blob/master/layer0/os_predef.h#L85

ericonr:
I'm not sure it's needed in those other headers, since we didn't get any other compilation error. But I don't think there's a reason not to simplify those includes, since including `<malloc.h>` and `<stddef.h>` at the same time shouldn't be an issue.

ericonr:
@speleo3 

What do you think about something like:

```diff
diff --git a/layer0/os_predef.h b/layer0/os_predef.h
index 7b1336f8..4aad41d7 100644
--- a/layer0/os_predef.h
+++ b/layer0/os_predef.h
@@ -82,11 +82,8 @@ Z* -------------------------------------------------------------------
 
 /* END PROPRIETARY CODE SEGMENT */
 
-#ifdef __linux__
 #include <malloc.h>
-#else
 #include <stddef.h>
-#endif
 
 #if defined(_MSC_VER)
 // conversion from '...' to '...', possible loss of data
diff --git a/ov/src/ov_port.h b/ov/src/ov_port.h
index 19a57a51..4fcc2eec 100644
--- a/ov/src/ov_port.h
+++ b/ov/src/ov_port.h
@@ -11,12 +11,8 @@
 #include <string.h>
 #include <stdlib.h>
 #include <ctype.h>
-
-#ifdef __linux__
 #include <malloc.h>
-#else
 #include <stddef.h>
-#endif
 
 #include "ov_defines.h"
 #include "ov_status.h"
```

speleo3:
I actually see no evidence that `malloc.h` would be needed on Linux, looks like a relict from the past. I think we can remove it.

Including `stddef.h` for all platforms sounds good.

ericonr:
@speleo3 updated!

speleo3:
Thank you!

PR #148, opened 2020-12-03
Support mmcif extension

keitaroyam:
e.g. Refmac output file

I did not test this, so please correct or discard it if this change is wrong. Thanks.


speleo3:
Thanks for the patch. We actually do not use the molfile pdbxplugin for CIF support, at least not by default.

The correct solution is to add an alias here: https://github.com/schrodinger/pymol-open-source/blob/6cb2b6f05fd8f45eb680c3bae7232648f8708f3b/modules/pymol/importing.py#L56-L97

keitaroyam:
Thank you very much for pointing this out! I will change it.

PR #177, opened 2021-04-28
Fixing SYBYL atom types for Sulphoxide sulphur (S.o) and Sulphone sulphur (S.o2)

bougui505:
The SYBYL atom typing for Sulphoxide sulphur (S.o) and Sulphone sulphur (S.o2) doesn't match the SYBYL nomenclature (https://www.sdsc.edu/CCMS/Packages/cambridge/pluto/atom_types.html)


JarrettSJohnson:
Looks good to me. Most other sources seem to agree with this, but one we've found has the 'o' upper-case: http://chemyang.ccnu.edu.cn/ccb/server/AIMMS/mol2.pdf

PR #180, opened 2021-05-18
Add desktop and metadata files.

kevinsmia1939:
I create desktop and metadata files. This is useful if any one want to create desktop icon or add a menu entry for PyMOL.


JarrettSJohnson:
Hi. Thanks for the PR. I'm a few years behind responding to this. However, we feel this is slightly out-of-scope of this repo and probably should be best left to downstream package maintainers instead.

PR #202, opened 2021-11-10
get_sasa_relative: Add subsele argument

speleo3:
This patch adds a `subsele` argument to the `get_sasa_relative` command. It allows me to query relative side chain exposure.

Example:
```
fetch 1ubq
get_sasa_relative polymer, subsele=sidechain
```

### DIFF
--- a/modules/pymol/util.py
+++ b/modules/pymol/util.py
@@ -1061,7 +1061,15 @@ DESCRIPTION
     _self.enable(name)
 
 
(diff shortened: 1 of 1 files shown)


JarrettSJohnson:
Looks good! Just gave it a quick test. Feel free to add a test to `pymol-testing` and a comment on PyMOL Wiki if you're free. If not, I'll add it when I do a quick round of those areas next week. :)

speleo3:
Thanks for merging! (what time is it? ;-)) I'll follow up soon with test and wiki.

PR #203, opened 2021-11-10
Add `fetch_path` to `is_session_blacklisted`

speleo3:
The `fetch_path` setting is usually not meaningful across different computers. Opening a session file which has `fetch_path` set to a location which doesn't exist on the current computer will make the `fetch` command fail to save files.


JarrettSJohnson:
Ouch. Thanks for this PR. I hope this hasn't caused you or anyone else too much trouble. Going to open an internal ticket to see if there are additional settings we should be on the look out for that should also be blacklisted.

speleo3:
Thanks for merging! It was a minor annoyance that caught me from time to time.

Issue #212, opened 2021-12-15
cmd.load_model discrete argument unexpected behavior


speleo3:
I would expect the following test to pass:

```python
from pymol import cmd
cmd.fragment("gly")
model = cmd.get_model()

cmd.load_model(model, "m1", discrete=0, state=1)
cmd.load_model(model, "m2", discrete=1, state=1)
cmd.load_model(model, "m3", discrete=-1, state=1)

assert cmd.count_discrete("m1") == 0
assert cmd.count_discrete("m2") == 1
assert cmd.count_discrete("m3") == 1

cmd.load_model(model, "m1", discrete=-1, state=2)
cmd.load_model(model, "m2", discrete=-1, state=2)
cmd.load_model(model, "m3", discrete=-1, state=2)

assert cmd.count_discrete("m1") == 0
assert cmd.count_discrete("m2") == 1
assert cmd.count_discrete("m3") == 1

cmd.load_model(model, "m1", discrete=0, state=3)
cmd.load_model(model, "m2", discrete=1, state=3)
cmd.load_model(model, "m3", discrete=1, state=3)

assert cmd.count_discrete("m1") == 0
assert cmd.count_discrete("m2") == 1
assert cmd.count_discrete("m3") == 1

cmd.load_model(model, "m1", discrete=1, state=4)
cmd.load_model(model, "m2", discrete=0, state=4)
cmd.load_model(model, "m3", discrete=0, state=4)

assert cmd.count_discrete("m1") == 1
assert cmd.count_discrete("m2") == 0
assert cmd.count_discrete("m3") == 0
```

speleo3:
I think this **pymol-users** message is related: [Loading chempy objects into PyMOL 2.4.0](https://sourceforge.net/p/pymol/mailman/message/37393799/)

PR #213, opened 2021-12-15
More consistent cmd.load_model discrete behavior

speleo3:
- Allow `discrete=0` for multi-state objects
- Keep existing discrete flag with `discrete=-1`

Fixes https://github.com/schrodinger/pymol-open-source/issues/212


JarrettSJohnson:
Thank you for the PR as always Mr. Holder!

Issue #225, opened 2022-03-01
Fallback MMTF loader support broken


speleo3:
If PyMOL is compiled with `--use-msgpackc=no` then it would [fall back](https://github.com/schrodinger/pymol-open-source/blob/f7048b42758fdc6117646eeffbb51432157ea82e/layer1/P.cpp#L1986-L1992) to `chempy.mmtf` to load MMTF files. This fallback is broken with [msgpack-python 1.0](https://github.com/msgpack/msgpack-python) which changed the default of `raw=True` to `raw=False`.

**Steps to reproduce:**

1. Compile PyMOL with `--use-msgpackc=no`
2. `fetch 1ubq, type=mmtf`

**Result:** Nothing gets loaded (and no error message)

PR #226, opened 2022-03-01
Fix fallback MMTF loader

speleo3:
Fixes `chempy.mmtf` with `msgpack-python` 1.0.

Also makes it work with `cmd.load_raw` and pass the corresponding tests.

Fixes https://github.com/schrodinger/pymol-open-source/issues/225


TstewDev:
Thanks for the pull request!

PR #233, opened 2022-04-22
Fix crash in find_molfile_plugin

speleo3:
Was calling Python function with unblocked interpreter (before `APIExit(G)`).

Caused a crash for on Linux with self-compiled PyMOL and trying to load an unsupported file format.


JarrettSJohnson:
Thank you!

PR #235, opened 2022-06-06
Use XDG_CONFIG_HOME for pymolrc

danpf:
Allows you to put your `.pymolrc` at `~/.config/pymol/pymolrc` to adhere to the XDG_CONFIG_HOME standard


JarrettSJohnson:
Thank you for the commit @danpf :)

PR #236, opened 2022-06-12
correct plugin install behavior

songyongshun:
When install a plugin, the .git directory under plugin should be ignored. 
Otherwise, a permission error will prompt when uninstalling the plugin, at least under Windows.


JarrettSJohnson:
Thanks for the commit! :)

PR #260, opened 2022-10-05
Fix build error in Pool.__del__

speleo3:
Python 3.10 doesn't like the global `multiprocessing.pool.ThreadPool` instance anymore.
```
Exception ignored in: <function Pool.__del__ at 0x7f5461251630>
Traceback (most recent call last):
  File ".../lib/python3.10/multiprocessing/pool.py", line 271, in __del__
  File ".../lib/python3.10/multiprocessing/queues.py", line 377, in put
  File ".../lib/python3.10/multiprocessing/connection.py", line 205, in send_bytes
  File ".../lib/python3.10/multiprocessing/connection.py", line 405, in _send_bytes
AttributeError: 'NoneType' object has no attribute 'pack'
```
This fix uses a local context manager instead.

### DIFF
--- a/monkeypatch_distutils.py
+++ b/monkeypatch_distutils.py
@@ -15,9 +15,10 @@ import distutils.unixccompiler
 from distutils import log
 from distutils.errors import (DistutilsExecError, DistutilsPlatformError,
                               CompileError, LibError, LinkError)
+from multiprocessing.pool import ThreadPool
 
 # threaded parallel map (optional)
-pmap = map
+parallel_jobs = 1
 
 try:
     import _osx_support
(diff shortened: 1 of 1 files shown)


JarrettSJohnson:
Thanks, Thomas. Internally, I've started to replace monkeypatch_distutils with a setuptools + cmake alternative due to the upcoming distutils obsolescence. Works fine on Windows, but I haven't finished it off yet due to a couple of incentive-only libraries being a bit tricky to be recognized by CMake on Mac/Linux. Off of the top of my head now, though, I don't think there's anything that is stopping it to work on open-source. I'll probably go ahead and give that a try with this repo soon.

speleo3:
Thanks Jarrett for the update. I'm curious to see the cmake solution at some point, that sounds exciting.

PR #261, opened 2022-10-05
Fix crash when loading discrete object

speleo3:
Fixes crash for
```
fetch 2xwu, type=cif, discrete=1
```
Follow-up on "PYMOL-3840 Fix cartoon rep memory init" 682572b9789cca856ea2be37bf707acdfb3e3bbb


JarrettSJohnson:
🎉

PR #262, opened 2022-10-06
Fix testSave_symmetry__mmtf for msgpack-cpp=4.1.2

speleo3:
Floats may be stored as integers in msgpack, and msgpack-cpp does that proactively since version 4.1.2. That broke the `unitCell` angle round trip in [testSave_symmetry__mmtf](https://github.com/schrodinger/pymol-testing/blob/ca649bf79409e92e7ae419d384dc610bffd118ca/tests/api/exporting.py#L281-L295).

See also:
https://github.com/rcsb/mmtf-c/pull/35
https://github.com/msgpack/msgpack-c/pull/1018


JarrettSJohnson:
Much appreciated as always.

PR #271, opened 2023-01-10
Workaround shader disable encountered in combination with wayland

bisdan:
Known workaround for glew initialization under wayland error that seems to be a glew build issue.


JarrettSJohnson:
Thanks for the PR!

PR #276, opened 2023-02-03
Avoid 3x2 grid for size=4

speleo3:
This pull request optimizes the number of columns and rows in grid mode. Previously, you could easily end up with a 3x2 instead of 2x2 grid with 4 elements, or a 4x2 instead of 3x2 grid for 5 elements.

Example:
```
set grid_mode
viewport 800, 400
fragment gly, m1
fragment gly, m2
fragment gly, m3
fragment gly, m4
```
| Current master        | This Pull Request  |
| ------------- |-----|
| grid-old | grid-new |

### DIFF
--- a/layer1/Scene.cpp
+++ b/layer1/Scene.cpp
@@ -194,6 +194,16 @@ void GridUpdate(GridInfo * I, float asp_ratio, GridMode mode, int size)
         else
           n_row++;
       }
+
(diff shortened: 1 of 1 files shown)


JarrettSJohnson:
Cool. Thanks!

speleo3:
Thank you Jarrett :)

PR #285, opened 2023-04-15
Fix GIL for get_object_ttt and friends

speleo3:
The following functions were releasing the GIL and then calling Python functions:

- CmdCombineObjectTTT
- CmdTransformObject
- CmdTransformSelection
- CmdGetObjectTTT

Fixed by either moving the Python calls outside the unblocked API context, or by using a blocked API context.

Observed as a crash after clicking the "Properties" dialog button, which calls `cmd.get_object_ttt`.

```
AddressSanitizer:DEADLYSIGNAL
=================================================================
==46349==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000010 (pc 0x7f09b5d253d3 bp 0x000000000010 sp 0x7ffed0cf7a80 T0)
==46349==The signal is caused by a READ memory access.
==46349==Hint: address points to the zero page.
    #0 0x7f09b5d253d3 in PyList_New (/usr/lib/libpython3.10.so.1.0+0x1253d3)
    #1 0x7f09b22aa62f in PConvFloatArrayToPyList(float const*, int, bool) layer1/PConv.cpp:956
    #2 0x7f09b290a691 in CmdGetObjectTTT(_object*, _object*) layer4/Cmd.cpp:4604
```

Likely also the reason for the crash reported in https://github.com/Pymol-Scripts/Pymol-script-repo/pull/129 which has a stack trace with `CmdTransformSelection`.

### DIFF
--- a/layer4/Cmd.cpp
+++ b/layer4/Cmd.cpp
@@ -1368,13 +1368,11 @@ static PyObject *CmdCombineObjectTTT(PyObject * self, PyObject * args)
   PyObject *m;
   float ttt[16];
   API_SETUP_ARGS(G, self, args, "OsO", &self, &name, &m);
+  if (PConvPyListToFloatArrayInPlace(m, ttt, 16) <= 0) {
+    return APIFailure(G, "Bad Matrix");
+  }
   API_ASSERT(APIEnterNotModal(G))
-  pymol::Result<> result;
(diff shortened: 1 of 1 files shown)


JarrettSJohnson:
Great catch. Thanks for the PR!

PR #301, opened 2023-07-28
Fix volume editor for Python 3.10

speleo3:
Related to https://github.com/Pymol-Scripts/Pymol-script-repo/issues/136


JarrettSJohnson:
Thanks Thomas. I'm going to be in and out of town for the next two-three weeks. When I come back, I can try helping out with some of the other plugins here too.

speleo3:
> I'm going to be in and out of town for the next two-three weeks. When I come back, I can try helping out with some of the other plugins here too.

Thanks Jarrett for letting me know.

Issue #322, opened 2023-12-19
get_atom_coords() fails with object or selection name


speleo3:
This fails:
```py
cmd.pseudoatom("foo")
print(cmd.get_atom_coords("foo"))
```
```
 Error: Empty Selection
```
Works if passing a selection expression which is not an exact match with an object/selection name:
```py
cmd.pseudoatom("foo")
print(cmd.get_atom_coords("(foo)"))
```
```
[0.0, 0.0, 0.0]
```
Reason seems to be the `SelectorTmp::getAtomCount()` function where `m_count == 0` for named selections.

https://github.com/schrodinger/pymol-open-source/blob/d59041b68ce95fc9f72c54a3cce72437fe679b3e/layer3/Selector.h#L236

speleo3:
No need to change anything immediately, I think this is a low priority bug (no crash and there is a workaround).

`SelectorTmp::getAtomCount` is only used once. You could consider removing it and calling `SelectorCountAtoms()` instead. Or call `SelectorCountAtoms()` in `SelectorTmp::getAtomCount` if `m_count == 0`. That would fix the issue.

Not creating a new temporary selection for a named selection makes sense. But the assumption that `SelectorGetTmp()` returns the atom count is not correct in that case.

PR #323, opened 2023-12-19
Fix SelectorTmp::getAtomCount() w/ selection name

speleo3:
Fall back to `SelectorCountAtoms()` if `m_count == 0`, which is the case for named selections.

Fixes https://github.com/schrodinger/pymol-open-source/issues/322


speleo3:
Thanks @TstewDev for merging!

PR #332, opened 2024-01-30
Fix for broken `png_check_sig` in libpng 1.6.41

speleo3:
See https://github.com/pnggroup/libpng/issues/534

The bug report says that `png_check_sig` was deprecated many years ago.


JarrettSJohnson:
Thanks. LGTM! I'll approve and merge this in when I'm done with the repo update.

PR #344, opened 2024-03-14
APBS electrostatics plugin

JarrettSJohnson:


speleo3 (approved):
Thank you @JarrettSJohnson for the review opportunity. I'm very excited that Schrödinger decided to open-source the plugin :partying_face:

Code looks good and works for me.

PR #350, opened 2024-03-23
Black formatting

pslacerda:
Free developer time from doing irrelevant formatting.


JarrettSJohnson:
Thanks for submitting this PR to format the python codebase. This is a goal that we're also working on for the upcoming year (probably within the next few weeks now since 3.0 is out the door), but it needs to also consider organization-wide agreements and guidelines. A lot of development time is wasted by unformatted/unreadable code (trust me, I battle with much of it on a daily basis :) ). I'll have to do this with a bit of consideration and preparation due to how we handle open-source and incentive--and both will have to be done in tandem.

pslacerda:
There is no AST change on black formatting. But it's fair to not black the fully source code on only one commit.

PR #362, opened 2024-05-05
replace distutils for python 3.12

branchvincent:
Python 3.12 has removed `distutils`: https://docs.python.org/3.12/whatsnew/3.12.html#removed

### DIFF
--- a/modules/pymol/plugins/installation.py
+++ b/modules/pymol/plugins/installation.py
@@ -45,8 +45,6 @@ def cmp_version(v1, v2):
     '''
     Compares two version strings. An empty version string is always considered
     smaller than a non-empty version string.
-
-    Uses distutils.version.StrictVersion to evaluate non-empty version strings.
     '''
     if v1 == v2:
         return 0
@@ -55,8 +53,9 @@ def cmp_version(v1, v2):
     if v2 == '':
         return 1
     try:
-        from distutils.version import StrictVersion as Version
-        return cmp(Version(v1), Version(v2))
+        v1_parts = list(map(int, v1.split('.')))
+        v2_parts = list(map(int, v2.split('.')))
+        return (v1_parts > v2_parts) - (v1_parts < v2_parts)
     except:
         print(' Warning: Version parsing failed for', v1, 'and/or', v2)
         return 0
--- a/create_shadertext.py
+++ b/create_shadertext.py
@@ -7,7 +7,6 @@ import glob
 from collections import defaultdict
 from os.path import dirname
 from subprocess import Popen, PIPE
-from distutils import dir_util
 
 def create_all(generated_dir, pymoldir="."):
     '''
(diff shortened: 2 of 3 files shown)


modules/pymol/plugins/installation.py
```diff
@@ -55,8 +53,7 @@ def cmp_version(v1, v2):
     if v2 == '':
         return 1
     try:
-        from distutils.version import StrictVersion as Version
-        return cmp(Version(v1), Version(v2))
+        return int(v1.split(".") < v2.split("."))
```
JarrettSJohnson, line 56: Fails assertions:
```python
assert cmp_version("3.1.0", "3.2.0") == -1
assert cmp_version("3.2.0", "3.2.0") == 0
assert cmp_version("3.5.0", "3.2.0") == 1
assert cmp_version("3.10.0", "3.2.0") == 1
```

I believe for a couple of reasons: the original `cmp` function returns -1, 0, 1 rather than 0/1. There's also a lex comparison before the integer cast, so "3.10.0" will be seen to be less than "3.2.0". On my local branch that fixes this I have:

```
        v1_parts = list(map(int, v1.split('.')))
        v2_parts = list(map(int, v2.split('.')))
        return (v1_parts > v2_parts) - (v1_parts < v2_parts)
```
which yields a list of versions expressed as a list of `int`s and compares them

JarrettSJohnson (changes requested):
Thanks for the PR. Just FYI, I'm currently also working on removing `distutils` (but haven't gotten around to it yet--need to make this a priority obviously ASAP). It's mostly due to the removal of `monkeypatch_distutils` which provides us parallel compilation and incremental builds; I just need it to get it working for Mac. I'll probably finish that up this weekend.

The changes proposed here seem fine, but just a suggestion below.

modules/pymol/plugins/installation.py
branchvincent: oops, my bad thanks for catching this!

JarrettSJohnson (approved):

branchvincent:
Thanks @JarrettSJohnson, btw another issue I've encountered is `setup.py` isn't compatible with `pip` (now preferred over the [deprecated](https://packaging.python.org/en/latest/discussions/setup-py-deprecated/) `python setup.py install`), which tries to build a wheel with a relative path for `install_libbase` (i.e. `build/bdist.macosx-14.0-arm64/wheel/`):

https://github.com/schrodinger/pymol-open-source/blob/6d9ddd4bd3eaf31b96b33a557b6f7c101bb2cebe/setup.py#L228

leading to

```console
$ cat $(which pymol)
#!/bin/sh
exec "/opt/homebrew/Cellar/pymol/3.0.0/libexec/bin/python" "/pymol/__init__.py" "$@"
```
any chance we could just use a [console script](https://setuptools.pypa.io/en/latest/userguide/entry_point.html#console-scripts) instead?

PR #369, opened 2024-06-10
Include cstdint header for CifFile module

speleo3:
I need `<cstdint>` on Arch Linux with gcc, otherwise `std::int8_t` is not defined.


JarrettSJohnson:
Thanks!

Issue #370, opened 2024-06-10
test_utils.requires_version("3.0") seems broken


speleo3:
The decorator `@test_utils.requires_version("3.0")` seems to be broken.

```sh
$ python -m pymol --version
PyMOL 3.1.0a0 Open-Source (dbf1ca58db), 2024-06-10
```
```sh
$ python -i tests/helpers/test_utils.py <<< 'print(compatible_with("3.0"))'
>>> False
>>> 
```
```sh
$ python -m pytest -v tests/api/test_importing.py 
Warning: use "from pymol import cmd" instead of "import cmd"
=========================================== test session starts ============================================
platform linux -- Python 3.12.3, pytest-8.2.1, pluggy-1.5.0 -- /d2/bld/env-open-source-3-1-master/bin/python
cachedir: .pytest_cache
rootdir: /d1/schrodinger/pymol-open-source/testing
configfile: pytest.ini
plugins: typeguard-4.2.1
collected 2 items

tests/api/test_importing.py::test_bcif SKIPPED (Requires PyMOL 3.0)                                   [ 50%]
tests/api/test_importing.py::test_bcif_array SKIPPED (Requires PyMOL 3.0)                             [100%]

============================================ 2 skipped in 0.61s ============================================

```

PR #372, opened 2024-06-10
Fix test_utils.compatible_with

speleo3:
Fixes https://github.com/schrodinger/pymol-open-source/issues/370

Solution is similar to https://github.com/schrodinger/pymol-open-source/blob/86ff2c3be6548c48f2e414e47036915ef3aaba7e/testing/testing.py#L87-L94


JarrettSJohnson:
Thanks. This uncovered some issues; I'll double check these on other machines.

PR #373, opened 2024-06-11
Fix DistSet SelectorGetDistSet function for H-bo…

norbert1000:
When using the dist command with mode=2 for structures that have sensibly placed polar hydrogens, it becomes apparent that many hydrogen bonds with very good geometry are missed. The problem only affects atom pairs in which both atoms can have a donor and an acceptor function (i.e. waters, Ser, Thr, His and ligands).

Origin of this behavior is the DistSet SelectorGetDistSet function in layer3/Selector.cpp:

```
if(ai1->hb_donor && ai2->hb_acceptor) {
//check for suitable geometry
} else if(ai1->hb_acceptor && ai2->hb_donor) {
//check for suitable geometry
} else {
//no hydrogen bond
}

```
If both atoms are donor and acceptor, both possibilities need to be tested.
@folmilsch will post a modification that takes care of this situation.


norbert1000:
Please delete this pull request, it is duplicated in #374. Sorry.

JarrettSJohnson:
Closed as requested

PR #375, opened 2024-06-11
adding building with pyproject.toml

ye11owSub:


JarrettSJohnson:
Thanks for your patience. Just returned today. I used your `pyproject.toml` file with a custom backend that forwarded arguments to `setup.py`. This seems to be a suggested approach from PyPA members; Pillow does this (I pretty much used their approach for this). Commit here: https://github.com/schrodinger/pymol-open-source/commit/ae5ff74657c4d7f247eae6dc17830fe3302db69d . Works for me locally on Windows with `pip install --config-settings testing=true .` to enable testing as you would before with `python setup.py build --testing`. I'll work on fixing some tests and CI that I broke along the way, but compiles and runs as expected.

ye11owSub:
Hi @JarrettSJohnson 
I hope you had a great vacation!
Regarding the PR, I have never tried to use a custom backend, thanks for the advice. So, If this option is suitable now, then it's super! 
I've added your changes to this one.

JarrettSJohnson:
Feel free to also refer to https://github.com/schrodinger/pymol-open-source/actions/runs/9860021376/workflow as it passed with all three operating systems.

ye11owSub:
Hi @JarrettSJohnson !

Could you please take a look at this PR and let me know if everything looks good to you? 
Also, I'm having trouble running the CI for this branch([link to CI](https://github.com/schrodinger/pymol-open-source/actions/runs/10271654736)). Can you please help me out with that?

JarrettSJohnson:
Accepted the CI run. Seems like for linux it's failing due to missing block:

```yaml
    - name: Pip upgrade and install build
      run: |
        pip install --upgrade pip
        pip install build
```

Generally looks good. Though can the functional changes be separated from the formatting changes in `setup.py` either by commit or PR?

setup.py
```diff
@@ -8,144 +8,310 @@
...
-    use_msgpackc = 'guess'
+    use_msgpackc = "guess"
     testing = False
     openvr = False
-    use_openmp = 'no' if MAC else 'yes'
-    use_vtkm = 'no'
-    vmd_plugins = True
+    use_openmp = "no" if MAC else "yes"
+    use_vtkm = "no"
+    vmd_plugins = False
```
JarrettSJohnson, line 196: This should be left at `True` by default for tests to pass (and so ofc by default users can use this to open several file types).

JarrettSJohnson (changes requested):

setup.py
ye11owSub: Oops, I didn't even notice when it changed. Fixed

ye11owSub:
@JarrettSJohnson I split the PR into smaller commits and ran the test locally. All should be good now.

JarrettSJohnson (approved):

JarrettSJohnson:
Thanks for the PR

PR #380, opened 2024-08-06
Add atom object name on the  `cmd.get_model()` returned model

pslacerda:
Now `cmd.get_model()` returns an atom model with the object name on it. I didn't want to use the `object` variable because it is a built-in, however I thought that any other name would be misleading.

### DIFF
--- a/layer2/CoordSet.h
+++ b/layer2/CoordSet.h
@@ -1,16 +1,16 @@
 
-/* 
+/*
 A* -------------------------------------------------------------------
 B* This file contains source code for the PyMOL computer program
-C* copyright 1998-2000 by Warren Lyford Delano of DeLano Scientific. 
+C* copyright 1998-2000 by Warren Lyford Delano of DeLano Scientific.
 D* -------------------------------------------------------------------
 E* It is unlawful to modify or remove this copyright notice.
 F* -------------------------------------------------------------------
-G* Please see the accompanying LICENSE file for further information. 
+G* Please see the accompanying LICENSE file for further information.
 H* -------------------------------------------------------------------
 I* Additional authors of this source file include:
--* 
--* 
+-*
+-*
 -*
 Z* -------------------------------------------------------------------
 */
@@ -122,7 +122,7 @@ struct CoordSet : CObjectState {
 
   pymol::vla<RefPosType> RefPos;
 
-  /* idea:  
+  /* idea:
      int start_atix, stop_atix <-- for discrete objects, we need
      something like this that would enable pymol to skip atoms not in the
      discrete state...question is: are these atoms sorted together right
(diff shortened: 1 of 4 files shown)


JarrettSJohnson:
Aside from the compilation errors, I'm wondering if this is the right approach. Wouldn't each atom's object name be duplicated (wasted space)? Would it make sense to have the name be stored per-object/model rather than per-atom?

pslacerda:
I had this doubt and thought that `resn` is duplicate all over atoms, specially for polymers and decided to replicate.

And I'm not sure if the same happens when the string is built from the C API, but Python's strings of the same value share the same memory space.

pslacerda:
I decided to return the object name on the atom model because it's convenient. It would be even more if it worked on `iterate`.

layer3/MoleculeExporter.cpp
```diff
@@ -1820,7 +1820,7 @@ struct MoleculeExporterChemPy : public MoleculeExporter {
 
   void writeAtom() override {
     PyObject *atom = CoordSetAtomToChemPyAtom(G,
-        m_iter.getAtomInfo(), m_coord, getRefPtr(),
+        m_iter.getAtomInfo(), m_last_obj, m_coord, getRefPtr(),
```
JarrettSJohnson, line 1823: I think `m_iter.obj` is meant to be used instead. `m_last_obj` I think is mostly just used to distinguish object separation from adjacent atoms being iterated.

layer2/CoordSet.h
```diff
@@ -202,8 +202,8 @@ bool CoordSetInsureOrthogonal(PyMOLGlobals * G,
     bool quiet=true);
 
 void CoordSetGetAverage(const CoordSet * I, float *v0);
-PyObject *CoordSetAtomToChemPyAtom(PyMOLGlobals * G, AtomInfoType * ai, const float *v,
-                                   const float *ref, int index, const double *matrix);
+PyObject *CoordSetAtomToChemPyAtom(PyMOLGlobals * G, AtomInfoType * ai, ObjectMolecule * obj,
+                                   const float *v, const float *ref, int index, const double *matrix);
```
JarrettSJohnson, line 206: Optional: `clang-format` used for formatting C++ files for changed lines.

In the future, there will be enforced format checks during CI runs, but I won't sweat over it for now.

JarrettSJohnson (changes requested):
LGTM, just mostly one suggestion regarding variable usage.

I take back my comment regarding having `model` on the `model` level since I was under the impression these were exported as separate models, but this probably makes the most sense to align w/ iterate.

JarrettSJohnson (approved):

JarrettSJohnson:
Thanks for the PR!

PR #391, opened 2024-08-17
refactoring shortcut.py script

ye11owSub:


ye11owSub:
@JarrettSJohnson Could you please take a look at this PR?

JarrettSJohnson:
Yup, I plan to review after 3.1 is out later this month.

ye11owSub:
> Yup, I plan to review after 3.1 is out later this month.

copy that, thank you

modules/pymol/shortcut.py
```diff
@@ -1,184 +1,171 @@
...
+        return keyword in self.shortcut
+
+    def __getitem__(self, keyword: str) -> Optional[int | str]:
+        return self.shortcut.get(keyword)
+
+    def __delitem__(self, keyword: str) -> None:
+        self.keywords.remove(keyword)
+        self.rebuild()
+
+    def make_abbreviation(self, s: str, groups_length: int) -> str:
```
JarrettSJohnson, line 51: Is this part of the public API (I didn't see any tests or usages in the codebase that uses this method explicitly)? If it's just meant to be internally used by `Shortcut`, can you add a leading underscore. I appreciate the examples here; also can you add a brief sentence or two on what this is achieving?

modules/pymol/shortcut.py
```diff
@@ -1,184 +1,171 @@
...
+            if keyword == abbreviation:
+                continue
+
+            self.abbreviation_dict[abbreviation].append(keyword)
+
+            for i in range(abbreviation.find("_") + 1, len(abbreviation)):
+                sub = abbreviation[0:i]
+                self.shortcut[sub] = 0 if sub in self.shortcut else keyword
+
+    def rebuild(self, keywords: Optional[Iterable] = None) -> None:
```
JarrettSJohnson, line 84: Can you add a docstring for this and the other methods as well?

tests/pymol/test_shortcut.py
JarrettSJohnson, line 1: Can this test be kept with the other tests? `test` and `testing` is already confusing as it is, and `tests` would just add to the confusion.

JarrettSJohnson (changes requested):

modules/pymol/shortcut.py
ye11owSub: done

modules/pymol/completing.py
```diff
@@ -1,14 +1,16 @@
...
     setting autocompletion.
     '''
-    def interpret(self, kee, mode=0):
-        if not kee.startswith('s.'):
-            return cmd.Shortcut.interpret(self, kee, mode)
-        v = cmd.setting.setting_sc.interpret(kee[2:])
+    def interpret(self, keyword, mode=False):
+        if not keyword.startswith('s.'):
+            return super().interpret(keyword, mode)
+        v = cmd.setting.setting_sc.interpret(keyword[2:])
```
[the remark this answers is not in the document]
ye11owSub: done

modules/pymol/shortcut.py
ye11owSub: done

tests/pymol/test_shortcut.py
ye11owSub: The tests in the `tests` folder are a duplicate of the tests in the `test` folder, except that the more modern and widely used pytests library is used instead. I would like to gradually replace the old tests without deleting them. What do you think about it ?

ye11owSub: it will be a long process, but in the end it should improve readability

JarrettSJohnson: We have some tests here that are `pytest` already (see https://github.com/schrodinger/pymol-open-source/blob/master/testing/tests/api/test_editing.py) for example. Some tests have been already converted to `pytest` from `unittest` in our other repo; I just haven't got around to bringing them over here.

ye11owSub: understood. I just replaced the old tests with new ones.

modules/pymol/shortcut.py
```diff
@@ -1,184 +1,203 @@
...
+
+        Example 2:
+        Input: s:'abc_def', groups_length: 2
+        Output: 'ab_def'
+        """
+        groups = s.split("_")
+        groups[:-1] = [c[0:groups_length] for c in groups[:-1]]
+        return "_".join(groups)
+
+    def optimize_symbols(self, keyword: str) -> None:
```
JarrettSJohnson, line 69: Same for this one; probably just for internal usage.

JarrettSJohnson (commented):
Looks good. Just this final thing and I'll merge it in later this week.

modules/pymol/shortcut.py
ye11owSub: done

JarrettSJohnson (approved):

JarrettSJohnson:
Thanks for your patience. Looks good.

PR #396, opened 2024-09-06
Fix list typehint

hp115:


JarrettSJohnson:
Thanks for the PR. Can I ask what the motivation is for the change? This is valid from Python 3.9, which is our minimum Python requirement https://github.com/schrodinger/pymol-open-source/blob/9d3061ca58d8b69d7dad74a68fc13fe81af0ff8e/INSTALL#L12

Also, Python 3.8 is scheduled to be EOL anyway in about a month.

ye11owSub:
I agree with @JarrettSJohnson.
@hp115 For more information, you can check PEP [585](https://peps.python.org/pep-0585/)

PR #400, opened 2024-09-24
Scoped enums for PyQt6 proper support

jrom99:
Hello, I've updated each Qt enum so that they work both with PyQt5 and PyQt6 without any major hacking.

I also updated the INSTALL file to reflect changes to `pip install .` since pip23 and pip24, where all arguments must be in a key=value form. I took the liberty to change how some optional dependencies can be set/unset to be more explicit (`--no-libxml` is now `--libxml=false`), and as far as I've checked, they are not used in the current github workflow.

On Linux, Pymol has a dash/dock gear icon and WMCLASS "python3", I've fixed this so now it exports the correct WMCLASS (which for gnome means that it is now able to find the correct icon).

I couldn't find how to set `PYMOL_PATH` in a way that `pip install .` is able to listen:

- Raises unknown argument error:
  - `pip install . --config-settings pymol-path=/some/path`
  - `pip install . --config-settings pymol_path=/some/path`
- Ignored (pymol is placed in `venvprefix/bin/pymol` or `pythonuserbase/bin/pymol`):
  - `PYMOL_PATH=/some/path/ pip install .`
  - `env PYMOL_PATH=/some/path/ pip install .`


ye11owSub:
LGTM

ye11owSub (approved):

JarrettSJohnson:
Thanks for putting work into this. By chance can you split the unrelated issues into their own respective PRs?

jrom99:
> Thanks for putting work into this. By chance can you split the unrelated issues into their own respective PRs?

Hello, I don't know how to split commits, should I close this PR and then create three new ones?

modules/pmg_qt/pymol_qt_gui.py
```diff
@@ -929,9 +932,9 @@ def command_set_cursor(self, i):
         return self.lineedit.setCursorPosition(i)
 
     def update_progress(self):
-        progress = int(self.cmd.get_progress() * 100)
+        progress = self.cmd.get_progress()
         if progress >= 0:
-            self.progressbar.setValue(progress)
+            self.progressbar.setValue(progress * 100)
```
JarrettSJohnson, line 937: Any reason for this change? Seems to undo the changes made in 11325a6014565700be9673236c9b7b6c856ea6d9

modules/pmg_qt/pymol_qt_gui.py
```diff
@@ -692,8 +695,8 @@ def update_pixels(*args):
             width = form.input_width_units.value()
             height = form.input_height_units.value()
             factor = get_factor()
-            form.input_width.setValue(int(width / factor))
-            form.input_height.setValue(int(height / factor))
+            form.input_width.setValue(width / factor)
+            form.input_height.setValue(height / factor)
```
JarrettSJohnson, line 699: Reason for revert? Seems to undo the fix for d6cb36410bb18727bc4fa1918608d9d6d354c62b

JarrettSJohnson:
> > Thanks for putting work into this. By chance can you split the unrelated issues into their own respective PRs?
> 
> Hello, I don't know how to split commits, should I close this PR and then create three new ones?

Several ways you can do it, but I tend to squash-merge PRs anyway. You can keep this one open for just the Qt changes (while taking out the other changes e.g. `git checkout master -- <file_path_to_revert>`) and putting those into a separate review.

modules/pmg_qt/pymol_qt_gui.py
jrom99: Those changes were in the branch qt6, from which I started. For now, I'll make a PR without them, then.

PR #404, opened 2024-10-20
Allow to override build date with SOURCE_DATE_EPOCH

bmwiedemann:
Allow to override build date with `SOURCE_DATE_EPOCH`
to make builds reproducible.
See https://reproducible-builds.org/ for why this is good and https://reproducible-builds.org/specs/source-date-epoch/ for the definition of this variable.


JarrettSJohnson:
Thank you for the link. This seems reasonable.

PR #418, opened 2025-01-10
Fix INSTALL instructions for newer versions of pip

jrom99:

I updated the INSTALL file to reflect changes to pip install . since pip23 and pip24, where all arguments must be in a key=value form. I took the liberty to change how some optional dependencies can be set/unset to be more explicit (--no-libxml is now --libxml=false), and as far as I've checked, they are not used in the current github workflow.

I couldn't find how to set PYMOL_PATH in a way that pip install . is able to listen:

- Raises unknown argument error:
        `pip install . --config-settings pymol-path=/some/path`
        `pip install . --config-settings pymol_path=/some/path`
- Ignored (pymol is placed in venvprefix/bin/pymol or pythonuserbase/bin/pymol):
        `PYMOL_PATH=/some/path/ pip install .`
        `env PYMOL_PATH=/some/path/ pip install .`

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -1,6 +1,6 @@
 
 ----------------------------------------------------------------------
-INSTALLATION VIA COMPILATION 
+INSTALLATION VIA COMPILATION
 ----------------------------------------------------------------------
 
 See also: http://pymolwiki.org/index.php/Linux_Install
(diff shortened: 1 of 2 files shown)


JarrettSJohnson:
Thanks. LGTM. I do prefer affirmative parse arg names like this. Seems more consistent and less error-prone.

PR #419, opened 2025-01-10
Fix unknown wmclass in linux (missing dock icon)

jrom99:

On Linux, Pymol has a dash/dock gear icon and WMCLASS "python3", I've fixed this so now it exports the correct WMCLASS (which for gnome means that it is now able to find the correct icon).

### DIFF
--- a/modules/pmg_qt/pymol_qt_gui.py
+++ b/modules/pmg_qt/pymol_qt_gui.py
@@ -1223,9 +1223,8 @@ def execapp():
     window = PyMOLQtGUI()
     window.setWindowTitle("PyMOL")
 
-    # fix linux dash icon/missing wmclass
-    QtCore.QCoreApplication.setApplicationName("PyMOL")
-    QtGui.QGuiApplication.setDesktopFileName("org.pymol.PyMOL")
+    # fix gnome/wayland dash icon/missing wmclass
+    app.setDesktopFileName("org.pymol.PyMOL")
 
     @commandoverloaddecorator
     def viewport(w=-1, h=-1, _self=None):
--- a/modules/pmg_qt/pymol_qt_gui.py
+++ b/modules/pmg_qt/pymol_qt_gui.py
@@ -1223,6 +1223,10 @@ def execapp():
     window = PyMOLQtGUI()
     window.setWindowTitle("PyMOL")
 
+    # fix linux dash icon/missing wmclass
+    QtCore.QCoreApplication.setApplicationName("PyMOL")
+    QtGui.QGuiApplication.setDesktopFileName("org.pymol.PyMOL")
+
     @commandoverloaddecorator
     def viewport(w=-1, h=-1, _self=None):
         window.viewportsignal.emit(int(w), int(h))


modules/pmg_qt/pymol_qt_gui.py
```diff
@@ -1223,6 +1223,10 @@ def execapp():
     window = PyMOLQtGUI()
     window.setWindowTitle("PyMOL")
 
+    # fix linux dash icon/missing wmclass
+    QtCore.QCoreApplication.setApplicationName("PyMOL")
```
speleo3, line 1227: For me, the application name is already `PyMOL`. Confirmed with:
```py
assert app.applicationName() == "PyMOL"
```

modules/pmg_qt/pymol_qt_gui.py
```diff
@@ -1223,6 +1223,10 @@ def execapp():
...
+    QtGui.QGuiApplication.setDesktopFileName("org.pymol.PyMOL")
```
speleo3, line 1228: why not use `app`?
```py
app.setDesktopFileName("org.pymol.PyMOL")
```

modules/pmg_qt/pymol_qt_gui.py
```diff
@@ -1223,6 +1223,10 @@ def execapp():
     window = PyMOLQtGUI()
     window.setWindowTitle("PyMOL")
 
+    # fix linux dash icon/missing wmclass
```
speleo3, line 1226: Doesn't seem to be a problem with KDE or IceWM on X11. So I suggest to be more specific that this is a fix for "Gnome Dash".

modules/pmg_qt/pymol_qt_gui.py
jrom99: It's been some time since I've read the docs on how to fix it, if I recall correctly it was because of transient pop ups without a parent.

modules/pmg_qt/pymol_qt_gui.py
jrom99: Do you have access to a system with KDE/Wayland? It seems that this is Wayland specific actually, since you can specify wmclass for xorg in the .desktop file directly.

speleo3: Under KDE/Wayland:
- Without `setDesktopFileName()`: Generic icon in taskbar ❌
- With `app.setDesktopFileName("org.pymol.PyMOL")` and launching from a `org.pymol.PyMOL.desktop` file with `Icon=application-x-clementine`: Taskbar shows `application-x-clementine` icon ✔️
- With `app.setDesktopFileName("org.pymol.PyMOL")` and launching from a differently named `.desktop` file (e.g. `PyMOL.desktop`): Generic icon in taskbar ❌

jrom99: Do pop ups also have these icons? Or is it only the main window?

speleo3: pop ups get the same icon.

speleo3 (approved):

JarrettSJohnson:
Thanks for your patience, and thanks @speleo3 for giving this a look.

PR #420, opened 2025-01-10
 Scoped enums for PyQt6 proper support

jrom99:

I've updated each Qt enum so that they work both with PyQt5 and PyQt6 without any major hacking.

Honestly, I don't know how either git or github works, so I tried setting the original repo as upstream and then applying a reset to not mess with the commit history. This erased my changes, so I re-did them manually[^1] before submitting this pull request, but it seems that some commits appeared on this pull request.

Since I only tested those changes after updating the install script, I'm not sure how I can test them again in this version.

Some notes/questions:
- On `modules/pmg_qt/TextEditor.py` line 90, `QMessageBox` seems to be unbound
- On `modules/pymol/plugins/managergui_qt.py` the functions `repo_changed` and `info_repo_plugin` freeze the GUI. I had some code to use a worker instead, should I put it in this PR?

[^1]: Except for the `addAction` change in modules/pmg_qt/pymol_qt_gui.py, which was raising some warnings on my editor
image


JarrettSJohnson:
Can you try to request a merge from `schrodinger::qt6` to `master`?

jrom99:
> Can you try to request a merge from to `master` instead of `schrodinger::qt6`? Should be a dropdown at the top when you click `Edit`.

My PC broke some time ago, so I'm on my phone for now. Is it still necessary?

Issue #425, opened 2025-01-28
Shortcut now missing `has_key` method


alexander-myronov:
## Describe the bug
After the refactoring of the Shortcut class in this PR (https://github.com/schrodinger/pymol-open-source/pull/391/files), it is now missing a `has_key` method, which is still expected in setting.py (https://github.com/schrodinger/pymol-open-source/blob/master/modules/pymol/setting.py#L91). 

## To Reproduce
```
pymol -c -d "set pdb_use_ter_records, 0"
```
causes
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/site-packages/pymol/setting.py", line 263, in set
    v = (type, _validate_value(type, value))
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/pymol/setting.py", line 91, in _validate_value
    if is_string(value) and boolean_sc.has_key(value):
                            ^^^^^^^^^^^^^^^^^^
AttributeError: 'Shortcut' object has no attribute 'has_key'
```

## Possible fix
As the new Shortcut implementation has a `__contains__` method, you can just do
```
if is_string(value) and value in boolean_sc:
```

## Environment
- **PyMOL Version:** PyMOL 3.1.0 Open-Source (78520103a9), 2025-01-27

ye11owSub:
@alexander-myronov 
Thank you for bringing this to attention. I didn't notice this during the refactoring

PR #426, opened 2025-01-28
removing the usage of has_key method of the Shortcut class

ye11owSub:
fixing [issue-425](https://github.com/schrodinger/pymol-open-source/issues/425)


JarrettSJohnson (approved):
Approved; reminding myself though that higher-level testing/coverage for stuff like this is sorely needed.

PR #439, opened 2025-03-04
Attempt to detect delocalized atoms

pslacerda:
Hi. This is my attempt to support the selection of delocalized atoms. I copy the following comment to turn potential limitations explicit.

```C++
      // simple rule which gets all aromatic C atoms right, but can
      // be wrong for example for neutral aromatic N atoms with degree 3 or
      // for neutral carboxy O atoms which PyMOL also assigns bond oder 4.
```

I want also to detect aromatic rings. Should I do proceed?

### DIFF
--- a/layer1/P.cpp
+++ b/layer1/P.cpp
@@ -615,9 +615,9 @@ static T const* get_member_pointer(S const* instance, size_t offset)
  * Should be equivalent to:
  *   OBAtom::GetExplicitValence() [Open Babel 3.0]
  */
-float getExplicitValence(ObjectMolecule const* obj, size_t atm)
+unsigned getExplicitValence(ObjectMolecule const* obj, size_t atm)
 {
-  float value = 0;
+  unsigned value = 0;
 
   for (auto const& item : AtomNeighbors(obj, atm)) {
     int const order = obj->Bond[item.bond].order;
--- a/layer1/P.cpp
+++ b/layer1/P.cpp
@@ -615,9 +615,9 @@ static T const* get_member_pointer(S const* instance, size_t offset)
  * Should be equivalent to:
  *   OBAtom::GetExplicitValence() [Open Babel 3.0]
  */
-static int getExplicitValence(ObjectMolecule const* obj, size_t atm)
+float getExplicitValence(ObjectMolecule const* obj, size_t atm)
 {
-  int value = 0;
+  float value = 0;
 
   for (auto const& item : AtomNeighbors(obj, atm)) {
     int const order = obj->Bond[item.bond].order;
@@ -643,7 +643,7 @@ static int getExplicitValence(ObjectMolecule const* obj, size_t atm)
  *   RDKit::Atom::getDegree()
  *   OBAtom::GetExplicitDegree() [Open Babel 3.0]
  */
-static unsigned getExplicitDegree(ObjectMolecule const* obj, size_t atm)
+unsigned getExplicitDegree(ObjectMolecule const* obj, size_t atm)
 {
   return AtomNeighbors(obj, atm).size();
 }
--- a/layer1/P.h
+++ b/layer1/P.h
@@ -56,7 +56,7 @@ Z* -------------------------------------------------------------------
 #define cPRunType_alter_state    2
 #define cPRunType_label          3
 
-float getExplicitValence(ObjectMolecule const* obj, size_t atm);
+unsigned getExplicitValence(ObjectMolecule const* obj, size_t atm);
 unsigned getExplicitDegree(ObjectMolecule const* obj, size_t atm);
 
 int PLabelExprUsesVariable(PyMOLGlobals * G, const char *expr, const char *var);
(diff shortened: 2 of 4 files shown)


pslacerda:
Currently I'm using [this workaround](https://github.com/pslacerda/XDrugPy/blob/e8402e094d180033dc491d9b1f1ed616f86770a5/xdrugpy/hotspots.py#L754).

layer1/P.cpp
```diff
@@ -615,9 +615,9 @@ static T const* get_member_pointer(S const* instance, size_t offset)
  * Should be equivalent to:
  *   OBAtom::GetExplicitValence() [Open Babel 3.0]
  */
-static int getExplicitValence(ObjectMolecule const* obj, size_t atm)
+float getExplicitValence(ObjectMolecule const* obj, size_t atm)
```
JarrettSJohnson, line 618: I think the type here should be left as int (or unsigned int); would be consistent with the corresponding functions from OpenBabel and RDKit which this function is based off of. Can you instead just cast either the numerator or denominator to float if you need a floating result below?

layer1/P.cpp
```diff
@@ -643,7 +643,7 @@ static int getExplicitValence(ObjectMolecule const* obj, size_t atm)
  *   RDKit::Atom::getDegree()
  *   OBAtom::GetExplicitDegree() [Open Babel 3.0]
  */
-static unsigned getExplicitDegree(ObjectMolecule const* obj, size_t atm)
+unsigned getExplicitDegree(ObjectMolecule const* obj, size_t atm)
```
JarrettSJohnson, line 646: I probably should have mentioned in the previous review, but this might be better served in the `ObjectMolecule` or `AtomInfo` file. (But I'll take care of that later).

JarrettSJohnson:
> I want also to detect aromatic rings. Should I do proceed?
> Currently I'm using [this workaround](https://github.com/pslacerda/XDrugPy/blob/e8402e094d180033dc491d9b1f1ed616f86770a5/xdrugpy/hotspots.py#L754).

Seems like that workaround might have some limitations. Unless it's already done, some work might need to put in for robust aromatic ring detection:

https://github.com/schrodinger/pymol-open-source/blob/17c6cbd96d52e9692fd298daec6c9bda273a8aad/layer2/Mol2Typing.cpp#L31-L46

layer1/P.cpp
pslacerda: I can try to move to `AtomInfo`;

layer1/P.cpp
pslacerda: The delocalization is indicated by the float number because otherwise it were a int, it would cast implicitly because the "+3/2" rule.

JarrettSJohnson: I haven't looked into this too much, but I thought the decision to make it `int` was intentional even with this rule. What makes PyMOL's case different that we have to return `float` here while the two APIs this is similar to (from `OpenBabel` and `RDKit`) still return `int`?

pslacerda:
Yest, it would enolve a lot more chemistry than I know. *Hückel's Rules* for aromaticity is the way to go.

But not forget the main issue here: found delocalized electrons, it can't be detected on the molecule because multiply delocalizations can make the valence/degree integer. What I'm propose is very simple: check if this number is "fractional" for the atom.

pslacerda:
OpenBabel simple count neighbors summing up the bond order.
https://github.com/openbabel/openbabel/blob/889c350feb179b43aa43985799910149d4eaa2bc/src/atom.cpp#L978

Note that what i'm proposing is very simple: divide the degree by the explicit valence.

JarrettSJohnson:
> OpenBabel simple count neighbors summing up the bond order. https://github.com/openbabel/openbabel/blob/889c350feb179b43aa43985799910149d4eaa2bc/src/atom.cpp#L978
> 
> Note that what i'm proposing is very simple: divide the degree by the explicit valence.

I'll check more in detail about how the other apps handle this. I already saw the implementations for OpenBabel but it's still unclear to me how this handles cases like this and how it's different than PyMOL.

To allay my concerns for now, how about
```cpp
float getExplicitValenceF(ObjectMolecule const* obj, size_t atm)
{
    // same implementation as before
}

unsigned getExplicitValence(ObjectMolecule const* obj, size_t atm)
{
    return static_cast<unsigned>(getExplicitValenceF(obj, atm));
}

bool AtomIsDelocalized(ObjectMolecule const* obj, size_t atm)
{
    auto val = getExplicitDegree(obj, atm) / getExplicitValenceF(obj, atm);
    return val != static_cast<int>(val);
}
```
?

pslacerda:
I did very much like you said.

JarrettSJohnson:
Thanks for PR

pslacerda:
I'm not sure if this deserve go into PyMOL, maybe works only on a few cases or valence and degree are correctly guessed?

pslacerda:
ping

PR #444, opened 2025-03-24
use Catch2 version 3.x

hsk17:
in 2022, Catch2 got a major update to v3.,slightly incompatible to v2.  recent gentoo, ubuntu, opensuse, fedora have v3.  while opensuse and fedora also provide a v2, gentoo and ubuntu 24.04 do not.  ubuntu 22.04 provides v2 only.
anyhow: this PR should have the pymol source changes neccessary to migrate to catch2 v3.

### DIFF
--- a/layerCTest/Test.cpp
+++ b/layerCTest/Test.cpp
@@ -8,7 +8,7 @@
 #include <fstream>
 #include <iostream>
 #define CATCH_CONFIG_RUNNER
-#include <catch2/catch.hpp>
+#include <catch2/catch_all.hpp>
 #include "Test.h"
 #include "TestCmdTest2.h"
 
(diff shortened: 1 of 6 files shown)


JarrettSJohnson:
Thanks for the PR. We have upgraded to V3 for our incentive repo, but there was a reason that we didn't make these changes for Open-Source yet. I need to double check why. Perhaps those reasons don't apply anymore.

JarrettSJohnson:
If you can get this to work on `Ubuntu 24.04`, I'd be happy to merge. I can get `conda-forge`s `Catch2` v3 to work on `24.04`, but getting the native package to link correctly doesn't seem to be as straightforward (see https://github.com/schrodinger/pymol-open-source/actions/runs/14284753956/job/40038475573 ).

Should just be the change here:
https://github.com/schrodinger/pymol-open-source/blob/c208c62c213ce4961941e8790ab301fb5d9e6011/.github/workflows/build.yml#L8

PR #449, opened 2025-04-14
Support cmd.iterate expression as a decorated function

pslacerda:


pslacerda:
It may break strange or unconventional uses of this API. I don't know, seems fairly enough.

JarrettSJohnson:
Can I ask what problem this is trying to solve?

I think the most convincing API I've seen for `iterate` is to directly convert the results to a list:

https://github.com/speleo3/pymol-psico/blob/ae2b92c262bca4ee2423b98993386a679c7ffb0d/tests/test_querying.py#L26 

or 

https://github.com/schrodinger/pymol-open-source/blob/c208c62c213ce4961941e8790ab301fb5d9e6011/modules/pymol/editor.py#L407

pslacerda:
Besides it being more pythonic like the inline lambda support, it also support any other case supported by an arbitrary function:

```py
my_dict = {}
c_a = []
@cmd.iterate("not elem H")
def one_more_iter(a):
    key = (a.model, a.index)
    my_dict[key] = a
    if a.name in ['CA']:
        c_a.append(key)
```

I's more generic for arbitrary cases.

PR #454, opened 2025-05-07
Fix integer overflow in hash function.

Augustin-Zidek:
Fix integer overflow in hash function causing ASAN failure.

Identical codegen: https://godbolt.org/z/Ed7W1h


JarrettSJohnson:
Thanks for the PR. Much appreciated.

PR #458, opened 2025-05-31
Eliminate two Biopython deprecation warnings

speleo3:
- `Bio.pairwise2` has been deprecated -> Use `Bio.Align.PairwiseAligner` instead
- `SeqRecord` constructor: Using a string as the sequence is deprecated -> Convert to `Seq`

The changes are based on
https://github.com/speleo3/pymol-psico/blob/ae2b92c262bc/psico/seqalign.py

Unit testing:
Covered by `tests/api/seqalign.py::TestSeqalign`

### DIFF
--- a/modules/pymol/seqalign.py
+++ b/modules/pymol/seqalign.py
@@ -11,6 +11,25 @@ License: BSD-2-Clause
 
 from pymol import cmd, CmdException
 
+import functools
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    import Bio.Align
+
+
+@functools.cache
+def _get_aligner_BLOSUM62() -> "Bio.Align.PairwiseAligner":
+    from Bio.Align import PairwiseAligner, substitution_matrices
+    blosum62 = substitution_matrices.load("BLOSUM62")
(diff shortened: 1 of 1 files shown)


ye11owSub:
🚀

modules/pymol/seqalign.py
```diff
@@ -11,6 +11,25 @@
...
+@functools.cache
+def _get_aligner_BLOSUM62() -> "Bio.Align.PairwiseAligner":
+    from Bio.Align import PairwiseAligner, substitution_matrices
+    blosum62 = substitution_matrices.load("BLOSUM62")
+    missing_codes = ''.join(set('JUO-.?').difference(blosum62.alphabet))
+    blosum62 = blosum62.select(blosum62.alphabet + missing_codes)
+    aligner = PairwiseAligner(internal_open_gap_score=-10,
+                              extend_gap_score=-.5,
+                              substitution_matrix=blosum62)
+    assert aligner.mode == "global"
```
JarrettSJohnson, line 30: Is the intent here that we should always use `global` even if Biopython changes this as its default? If so, would a
```python
aligner.mode = "global"
```
be more future-proof?

JarrettSJohnson (approved):
LGTM just a quick Q

PR #480, opened 2025-11-03
Add support for >=, <= and == operators in selections

### LINKED ISSUE #477: Selection algebra for properties and enums
**Is your feature request related to a problem? Please describe.**
The selection algebra doesn't support `>=` and `<=` on property selectors (e.g. `select * & p.PROP <= 10`). A very related issue is that the equality operator doesn't exist (?) so we can't compare enums (integers) or strings.

**Describe the solution you'd like**
The solution I'd like would be `* & p.P1>=500 & p.P2=="MyFilter"` with `==` a synonymous to `=`. 

I'd like to add support to this, but don't know if I know enough for it.

pslacerda1:
Fix #477.

_Edit:_ in love with claude.ai

### DIFF
--- a/layer3/Selector.cpp
+++ b/layer3/Selector.cpp
@@ -8504,169 +8504,49 @@ static int SelectorSelect2(PyMOLGlobals * G, EvalElem * base, int state)
   case SELE_BVLx:
   case SELE_QVLx:
     oper = WordKey(G, AtOper, base[1].text(), 4, ignore_case, &exact);
-    if(!oper)
+    if(!oper) {
       ok = ErrMessage(G, "Selector", "Invalid Operator.");
-    if(ok) {
-      switch (oper) {
-      case SCMP_GTHN:
-      case SCMP_LTHN:
-      case SCMP_EQAL:
-        if(sscanf(base[2].text(), "%f", &comp1) != 1)
-          ok = ErrMessage(G, "Selector", "Invalid Number");
-        break;
+      break;
+    }
+    switch (oper) {
+    case SCMP_GTHN:
+    case SCMP_LTHN:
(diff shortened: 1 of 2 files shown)


JarrettSJohnson:
Just FYI, I am happy with the spirit of the PR, but just note that we're about to transition over to the next version. And since this has such a large scope-of-effect since this touches fundamental selection logic, this will be deferred until after we release (or branch for) 3.2.

pslacerda1:
I rather prefer to increase the test coverage in this area than revisit code that is already done. May this be an option?

JarrettSJohnson:
I didn't know this blocked the other PR. But I think either way, most of this PR still touches too much selection code--even if you were to add additional tests, who knows what else could regress that we haven't considered. IMO, this will probably have to wait until the next cycle.

JarrettSJohnson:
Support for C `PyMOLGlobals`-dependent tests are still not available (passing `NULL` to `SelectorParse` will try to dereference it when `Feedback` is called--so just avoid doing this). So feel free to just stick with the Python tests for now. The PR can contain both commits if you want, btw. I just wont squash them when I merge.

pslacerda1:
I was since the morning fighting with the code including a syntax error (unbalanced parenthesis) and other "IDE-time" errors my environment didn't catch. I'm lucky to have Gemini that solved my problems right in.

pslacerda1:
Hi @JarrettSJohnson. I see that you open-sourced the properties feature. Great!

But now, this feature isn't running on Incentive 3.1.8.

JarrettSJohnson:
Because this was released after 3.1, and we only do bug/fix patches for 3.1. This will be available in incentive 3.2.

PR #481, opened 2025-11-04
Validate selections in `get_object_list()``

### LINKED ISSUE #478: Sometimes `get_object_list(...)` returns `None` instead of `[]` or raising an exception
Makes sense to change `cmd.get_object_list(...)` to return empty lists instead of `None` in case of empty selections?
```py
def get_object_list(selection="(all)", quiet=1, *, _self=cmd):
        '''
        
DESCRIPTION

    "get_object_list" is an unsupported command that may have
    something to do with querying the objects covered by a selection.
    '''
        selection = selector.process(selection)
        with _self.lockcm:
            r = _cmd.get_object_list(_self._COb,str(selection))
            r = r or []
            if not quiet:
                print(" get_object_list: ",str(r))
        return r
```

pslacerda1:
Raise exceptions for empty or invalid inputs.

May fix #478.

### DIFF
--- a/modules/pymol/querying.py
+++ b/modules/pymol/querying.py
@@ -137,8 +137,12 @@ DESCRIPTION
     something to do with querying the objects covered by a selection.
     '''
         selection = selector.process(selection)
+        if selection.strip() == "":
+            raise pymol.CmdException("Empty selections or containing only spaces are invalid.")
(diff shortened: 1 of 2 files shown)


pslacerda1:
Yes, it mostly suits my case.

PR #485, opened 2025-11-29
Some new_command fixes

pslacerda1:


pslacerda1:
I started a wiki page for [`new_command`](https://pymolwiki.org/index.php/New_Command) and think that everything is ok by now.

Can you point me on how to make the community aware so they can also test?

JarrettSJohnson:
Hi Pedro, couple of high-level things first:

1. For your PRs, can you make a description of a summary/what the PR is trying to accomplish?
2. Can I ask why you're removing `helping.py` in `tests`?
3. Changes to the `.gitignore` would be better suited as a separate PR.

JarrettSJohnson:
How are you running the tests?
```
pymol -ckqy testing/testing.py --run all
```
like in the CI, I can reproduce the issue locally as well with this branch.

pslacerda1:
Hi @JarrettSJohnson, do you have any other questions?

I'm still unsure on how to promote this feature into the community. However I think that it unlock a lot, like better generated documentation and maybe even better command line features like auto-completion.

JarrettSJohnson:
I'll take most of the fault here, but I probably should have better anticipated how much complexity would be involved for a project like this. I'm definitely in favor of high-level goals here, but I wonder if we should have spent more time in the beginning with a clearer design doc and possible implementation off-repo (ideally in a branch) that we could iterate through before jumping into an implementation that's merged into master.

One of the things that I wonder is if a lot of this type checking could have been assisted by libraries like `argparse` or `click`. These are problems that have been pondered on by experts for a long time, and I worry that there's too much bespoke code here for this to be robust enough for everyone's use-case. Moreover, it's hard for me to review myself, because it feels like I am reviewing "Python typing" code and not "PyMOL" code. A lot of this is out of my wheelhouse and the devs are left to be the maintainers of this code going forward. The advantages of `new_command` are nice, but the maintenance and complexity is a significant burden (and may even outweigh the benefits) for me especially if it's meant to have the same reach as `extend`.

IMO, I think at this point, one of the best ways forward is perhaps to mark this as experimental for this upcoming release and reassess afterward--during which we can test it out on some simple plugins. Alternatively we could move `new_command` to a separate branch and experiment with it there.

As far as advertisement, I think PyMOLWiki is sufficient; though you may need to be your best advocate for something like this and test it on your plugins so we can see observe its strengths.

I'd like to cc the Thomas's @TstewDev and @speleo3 if they have any additional high-level comments or concerns. There's some details in the code that I can point for improvements, but I can follow up on those after once we align on a direction.

pslacerda1:
Be tranquility.

I did a lot of laboratory/testing before myself (e.g. parsing the abstract source tree and PyMOL style docstrings) and opted by this design and like the current overall design even if it's also a bit more complicated then I wanted.

Very much like `extend` it is designed to be developed for a time span and then be done... and forget. My current plan include innovative documentation declaration on comments and beautiful help visualization for commands that opt in (pretty much only and simple pretty print the formatted function/command signature). Blend source code, function declaration and command options in a single monolithic piece which is the signature is awesome in my opinion.

pslacerda1:
Take a look on the wiki: https://pymolwiki.org/index.php/New_Command. My specification is a `Runnable`.

pslacerda1:
Click is nice, they have some nice features like `File`, `Path`, `IntRange` and `FloatRange`, the remaining types I already got in current implementation; for instance, `click.Choice` can be an `Enum` or `StrEnum`. Anyway, `File` is just an convenience function over `click.Path` and `click.Path` seems somewhat a convenience of `pathlib.Path`.

```py
@click.command()
@click.argument('input', type=click.File('rb'))
@click.argument('output', type=click.File('wb'))
def inout(input, output):
    """Copy contents of INPUT to OUTPUT."""
    while True:
        chunk = input.read(1024)
        if not chunk:
            break
        output.write(chunk)
```

```py
@click.command()
@click.argument('filename', type=click.Path(exists=True))
def touch(filename):
    """Print FILENAME if the file exists."""
    click.echo(click.format_filename(filename))
```

pslacerda1:
These are components every human-computer interface has to handle case-by-case, while nobody invented something really abstract about this, we'll continuing to re-inventing the interfaces: click, Qt, Jupyter Widgets, Shiny... because of this I think we should focus on PyMOL specific stuff.

pslacerda1:
> IMO, I think at this point, one of the best ways forward is perhaps to mark this as experimental for this upcoming release and reassess afterward--during which we can test it out on some simple plugins.

Experimental support would be nice for experimental and toy projects.

My dreams are big, but are achievable. Like `@pm.extend` has been untouched by 14 years according to git blame, I think that `@pm.new_command` can also become finished and forgotten. The scope is not so big, in order: parsing, help and auto-completion. For now, I want to think only about parsing.

Common guy, Is very harder to draw these cartoons or develop such versatile selection engine.

Today I read about BearType and Typeguard, they're tools for type checking at runtime based on annotated type hints and are analog to what mypy does statically. They can be disabled any time to get full speed. Along with PEP 593 and its standard library implementation `typing.Annotated[<type>, <metadata>]`,  particularly if metadata has assertive predicates (like `Is[ ]` validator from BearType), we can get wonderful results: defensive programming just by declaring types.

(No, I don't think someone would invent a thing that would replace both Shiny and Click at the same time!)

--------------------------------
**Edit:**

I did as you told and moved the implementation to a better place. I think maybe the subset I covered it's ready. Currently I'm thinking about the improved annotation support in PyMOL 3.14 through [annotationlib](https://docs.python.org/3/library/annotationlib.html#module-annotationlib).

https://github.com/pslacerda1/pymol_new_command

What can I do gain your attention to this necessary feature?
@JarrettSJohnson @speleo3

JarrettSJohnson:
Hi Pedro, I need to dedicate a solid amount of time to think and address this carefully, but there's just a lot that I'm juggling with in regards to other projects. I promise I haven't forgotten. We still have this as a possibility for discussion to be included in 3.2 (and other features), but we haven't had that discussion yet.

pslacerda1:
There is some overlap between `parsing.prepare_call` and my function, they should share a common code.

JarrettSJohnson:
Thanks, and sorry for the hold-up. I still haven't had time to dedicate toward this, but to get the ball rolling towards wrapping up 3.2, I'll merge this and mark as experimental later so that it's at least available for you to use.

pslacerda1:
@JarrettSJohnson, besides `@new_command` decorator, I provide the `new_help` command at https://github.com/pslacerda1/pymol_new_command. It displays beautifully formatted help for commands declared with `@new_command`. The current implementation has some quirks like no special help for positional only arguments or out-of-order terminal logging, for instance.

The only code needed to change is the `help` function:
```py
    ...
    if hasattr(_self.keyword[cmmd][0], 'func'):
        return new_help(cmd_name=cmmd)
    ...
```

The remaining is new code introduced with my changes. Maybe can you or someone test it and help me review the features of `@new_command` and `new_help`.


image

PR #499, opened 2026-02-24
Refactor list_scan to handle new pymol-scripts-repo layout

jrom99:
https://github.com/Pymol-Scripts/Pymol-script-repo has changed its layout to split plugins into scripts, plugins and modules, and PyMOL can't find the plugins in their new locations.

### DIFF
--- a/modules/pymol/plugins/repository.py
+++ b/modules/pymol/plugins/repository.py
@@ -195,7 +195,7 @@ class GithubRepository(HttpRepository):
 
         blobs = []
         for d in r["tree"]:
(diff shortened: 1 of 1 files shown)


JarrettSJohnson:
Thanks for the PR. This is another item I forgot to port over to open-source. The major difference between our diffs is that you're also adding `modules` and `plugins` which may not have been the behavior before? Perhaps an improvement, but I'll test it out first.

JarrettSJohnson (changes requested):

JarrettSJohnson (approved):

